PR #12, opened 2019-09-08
Both programs: pascal and k diff committed. LC verified

manishjangid:


Problem1.c
```diff
@@ -0,0 +1,102 @@
...
+     [1],
+    [1,1],
+   [1,2,1],
+  [1,3,3,1],
+ [1,4,6,4,1]
+]
+
+
+ *
+ * Time Complexity : O(n^2) worst case
```
kaurramanpreet: why did you mention the worst case
isnt this for every case equal to n*(n+1)/2 which is equal to O(N)

manishjangid: Hi @kaurramanpreet , Yes, it's wrong it should be O(N). Here I didn;t take into account that only the last iteration of the loop will be running for N entries. it's a progressive one. I will fix it.
Thanks for pointing it out

kaurramanpreet:
No manish what i said was there is no worst case in this question every case is the same
So time complexity is not actually o(n) its o(n*n) only as o(n*(n+1)) is actually equal to n2 only but the point is it is same for all cases there is no worst or best case here

manishjangid:
Sure. Thanks for the explanation.

PR #19, opened 2020-01-13
all problems complete

nikrasiya:


532_k_diff_pairs_in_an_array.py
```diff
@@ -0,0 +1,71 @@
...
+            if (k > 0 and key - k in dic) or (k == 0 and dic[key] > 1):
+                result += 1
+        return result
+
+    def find_pairs_two_pointers(self, nums: List[int], k: int) -> int:
+        """
+          // Time Complexity :
+                            O(nlogn) Sorting
+          // Space Complexity : O(n)
+          // Did this code successfully run on Leetcode : Yes
```
super30admin: Why O(n)?

118_pascal's_triangle.py
```diff
@@ -0,0 +1,43 @@
...
+        #     for j in range(1, i):
+        #         result[i][j] = result[i - 1][j - 1] + result[i - 1][j]
+        # return result
+
+        # Approach 2
+        """
+          // Time Complexity :  O(n^2)
+                            'n' is the number of rows
+          // Space Complexity : O(n^2)
+          // Did this code successfully run on Leetcode : Yes
```
super30admin: Why O(n2)? space

532_k_diff_pairs_in_an_array.py
nikrasiya: Using sort function which internally uses Tim sort.

118_pascal's_triangle.py
nikrasiya: For each value of 'n', we have corresponding number of elements. 
For n =5, 1+2+3+4+5 = (n * n+1)/2 = n^2. No?

532_k_diff_pairs_in_an_array.py
nikrasiya: Please correct me

PR #24, opened 2020-02-26
Solution - Competitive Coding 3

saumilsjsu:


super30admin:
The logic is incorrect as how you are moving your pointers. Here's a better solution

PR #32, opened 2020-04-06
Pair with diff k cmplted

schakrad:


devashishnyati:
Good job Sindhu!
Submit other problems as well.

PR #39, opened 2020-04-07
Completed Comptetive Coding : Please review, referenced the Pascal solution

KBangera:


Pascals.java
```diff
@@ -0,0 +1,38 @@
+// Time Complexity :O(n^2)
+// Space Complexity : O(n)
```
devashishnyati, line 2: Why is space O(n)?

devashishnyati:
Good job Karthik!
You can optimize the find pairs to O(n) time by using extra space.

PR #40, opened 2020-04-07
Create pascals_triangle.py

megrao:


pascals_triangle.py
```diff
@@ -0,0 +1,41 @@
+Time and space complexity = O(NumRows^2)
```
devashishnyati, line 1: Why is space O(n^2)?

devashishnyati:
Good job Megha!
Submit other problem as well.

pascals_triangle.py
megrao: If number of rows is 5, we will need a list of lists of size 5*5 to provide the result, unless we are not counting it?

PR #49, opened 2020-05-11
Solved

tejassjain:


pascals_triangle.java
```diff
@@ -0,0 +1,30 @@
+//Time Complexity: O(numRows^2);
+//Space Complexity: O(numRows^2);
```
sowmyadvn, line 2: Is list considered as auxiliary space here?

sowmyadvn:
Please review space complexity of Pascal's triangle problem Tejas. Good job with the solutions!

sowmyadvn:
Please review space complexity of Pascal's triangle problem Tejas. Good job with the solutions!

pascals_triangle.java
tejassjain: We are taking an ArrayList of lists so it should he N^2 (that's what I think). What should it be?

sowmyadvn: Because we are returning the list as result, it is not considered as auxiliary space. So, space would be O(1) here.  If you use an intermediate data structure and don't return it, you have to consider it as additional space.

tejassjain: But aren't we calculating all the elements of the resultant list till the input number? So we are actually creating an ArrayList of lists. So that way it becomes O(N^2)

sowmyadvn: Let me ping you on Github so we can discuss in detail

PR #53, opened 2020-05-12
Completed Design

JacobStephen9999:


Kdiff pairs.py
```diff
@@ -0,0 +1,28 @@
+#Time Complexity : O(Nlogn) for sorting and O(N) for traversing the nums array
```
iivaishnavii, line 1: Can we do it with a better time complexity?

iivaishnavii:
Good job, Jacob!

PR #72, opened 2020-06-15
Competitive coding 3 : Please review it

nareshkumar-ss:


AbhiniveshP:
For Pascal's triangle question, you are returning ArrayList al and apart from that you have used one more temporary ArrayList al1. So, accordingly, the extra space (not including al) would be O(n) and not O(1).
For k-diff pairs question, try implementing HashMap approach as it is the optimal one in terms of Time Complexity

AbhiniveshP:
For Pascal's triangle question, you are returning ArrayList al and apart from that you have used one more temporary ArrayList al1. So, accordingly, the extra space (not including al) would be O(n) and not O(1).
For k-diff pairs question, try implementing HashMap approach as it is the optimal one in terms of Time Complexity

PR #76, opened 2020-06-15
Competitive_Coding-3 Done!

sarthaksugandhi9:


AbhiniveshP:
Nice work Sarthak !!!
In Pascal's triangle question, apart from the ArrayList that you are returning (res), you have used one more temporary ArrayList row, which would make the Space Complexity O(n).
In k-diff pairs question, try implementing HashMap approach as it is optimal in terms of Time Complexity.

AbhiniveshP:
Nice work Sarthak !!!
In Pascal's triangle question, apart from the ArrayList that you are returning (res), you have used one more temporary ArrayList row, which would make the Space Complexity O(n).
In k-diff pairs question, try implementing HashMap approach as it is optimal in terms of Time Complexity.

PR #80, opened 2020-06-15
Completed both the problems, please review.

samarthjj:


AbhiniveshP:
I think you have implemented a different Pascal's triangle variant. You need to return only the last row and not the entire matrix.
Also, in k-diff pairs question, you have sorted the nums array and so Time Complexity would be O(nlogn) and not O(n). For this question, also try implementing the HashMap solution as it is the optimal one in terms of Time Complexity.

AbhiniveshP:
I think you have implemented a different Pascal's triangle variant. You need to return only the last row and not the entire matrix.
Also, in k-diff pairs question, you have sorted the nums array and so Time Complexity would be O(nlogn) and not O(n). For this question, also try implementing the HashMap solution as it is the optimal one in terms of Time Complexity.

PR #97, opened 2020-06-23
Add files via upload

trivedipratul:


PA_43.py
```diff
@@ -0,0 +1,31 @@
+#LC 532 - K-diff Pairs in an Array
+#Time complexity - O(nlogn)
```
Lohith-Aswathappa, line 2: Try to optimize the code even more

Lohith-Aswathappa:
Great work, Pratul!
Review the inline comment.

PR #129, opened 2020-09-08
Add files via upload

kjayashr:
* This was asked during my interview.
* This problem works on my LC


Sumesh-H (commented):
Good work. Krishna! Kindly mention TC & SC of the problem. Also, try the other problem as well.

Sumesh-H:
Good work. Krishna! Kindly mention TC & SC of the problem. Also, try the other problem as well.

PR #134, opened 2020-09-09
Completed Competitive_Coding-3 - Please review

gssanjana4:


khushal1198 (commented):
Good work Sanjana. For the pascal triangle problem, you needed to return the whole triangle, not just the last row.

khushal1198:
Good work Sanjana. For the pascal triangle problem, you needed to return the whole triangle, not just the last row.

gssanjana4:
I thought the question is Pascals Triangle 2, which asks us to send kth row in Pascal's triangle with O(k) Time complexity

PR #141, opened 2020-09-11
Please review this submission

sidmal11:


samarthjj:
Good work Siddhart, Please also try **K-diff Pairs** problem using two pointers as well.

samarthjj:
Good work Siddhart, Please also try **K-diff Pairs** problem using two pointers as well.

PR #174, opened 2020-11-06
Completed both the problems

kateganesh:


samarthjj:
Good work here Ganesh, Good work on the comments.

samarthjj:
Good work here Ganesh, Good work on the comments.

PR #191, opened 2020-12-12
Add Solution -  Please Review

hvp004:


Himanshu-Agarwal93 (commented):
You could do better in k-diff pair with a binary search.

Himanshu-Agarwal93:
You could do better in k-diff pair with a binary search.

PR #195, opened 2020-12-15
Competitive_Coding-3 : Java implementation

ckallianpur:


super30admin (commented):
There is O(n) solution for K diff pairs 

`public class Solution {
    public int findPairs(int[] nums, int k) {

        int result = 0;

        HashMap <Integer,Integer> counter = new HashMap<>();
        for (int n: nums) {
            counter.put(n, counter.getOrDefault(n, 0)+1);
        }


        for (Map.Entry <Integer, Integer> entry: counter.entrySet()) {
            int x = entry.getKey();
            int val = entry.getValue();
            if (k > 0 && counter.containsKey(x + k)) {
                result++;
            } else if (k == 0 && val > 1) {
                result++;
            }
        }
        return result;
    }
}`

super30admin:
Please review the above comment. Learn O(N) solution of k diff pair problem

PR #244, opened 2021-02-07
Implemented Longest Palindrome - Please review

rachana-code:
Works on most test cases - on long one - time limit exceeded. Please suggest improvements.

https://leetcode.com/problems/longest-palindromic-substring/submissions/

### DIFF
--- a/longestPalindrome.java
+++ b/longestPalindrome.java
@@ -1,0 +1,69 @@
+class Solution {
+        public String longestPalindrome(String s) {
+            String result = s.substring(0, 1);
+            int len = s.length();
+            int left = 0;
+            int right = 0;
+
+            for(int i=0;i<len;i++){
+                left = i - 1;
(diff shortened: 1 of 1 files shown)


kannan-mrugapphan (commented):
This is a complicated version of the n^2 solution. I would recommend you to follow a sliding window approach where you continuously expand the window as long as the snippet is valid and shrink otherwise. This makes sure that each character is visited at a max of 2 times to get to a run time of n. Please reach out through a DM and I can walk you through this if needed.

kannan-mrugapphan:
This is a complicated version of the n^2 solution. I would recommend you to follow a sliding window approach where you continuously expand the window as long as the snippet is valid and shrink otherwise. This makes sure that each character is visited at a max of 2 times to get to a run time of n. Please reach out through a DM and I can walk you through this if needed.

PR #267, opened 2021-03-12
Finished Competitive Coding 3, please review!

shrns30:


khushal1198 (commented):
Good. Please submit the second problem of competitive coding as well!

khushal1198:
Good. Please submit the second problem of competitive coding as well!

PR #280, opened 2021-04-11
Please Review: Completed Competitive-Coding-3

NaisargiP:


khushal1198 (commented):
Good work with Pascal Triangle. Please add the second problem as well!

khushal1198:
Good work with Pascal Triangle. Please add the second problem as well!

PR #392, opened 2021-08-19
Competitive_Coding-3, Pls review

ysrinu1:


Rooppesh:
Good Job! Please add the TC, SC and Approach used and complete the other problem as well!

Rooppesh:
Good Job! Please add the TC, SC and Approach used and complete the other problem as well!

PR #410, opened 2021-09-10
pascals-triangle.py

ami-doshi:


maithrreye:
Good Job Ami ! please add the other problem too.

maithrreye:
Good Job Ami ! please add the other problem too.

PR #469, opened 2022-01-11
python solution

gdg4dev:


sonalgupta276 (commented):
Good job! Please check TC for find pairs

PR #481, opened 2022-01-17
Create problem1.py

rucha2997:


sonalgupta276 (commented):
Complete other problem as well. Mention TC and SC

sonalgupta276:
Complete other problem as well. Mention TC and SC

PR #508, opened 2022-02-28
Completed Competitive Coding-3 : Please Review

aryyamajana:


Problem2.java
```diff
@@ -0,0 +1,112 @@
+// Time Complexity : O(n2) where n = number of nodes in the tree
```
shrivatsav, line 1: Please check the time complexity.

shrivatsav (commented):
Good Work!

shrivatsav:
Good Work!

Problem2.java
aryyamajana: I know the time complexity will be O(nh) where n = number of nodes and h = height of tree. In average case, for a fully balanced tree, height will be log n. So it should be O(nlog n). 
But if I consider a skewed tree (skewed on one side) and disregard the early stopping as it will return false before traversing the other nodes, then the worst case time complexity will be O(n2).
So, which one should I consider? Could you please let me know your feedback and if my thought process is incorrect?

shrivatsav: >I know the time complexity will be O(nh) where n = number of nodes and h = height of tree. In average case, for a fully balanced tree, height will be log n. So it should be O(nlog n).


Correct. 

> But if I consider a skewed tree (skewed on one side) and disregard the early stopping as it will return false before traversing the other nodes, then the worst case time complexity will be O(n2).

If it is a skewed Binary Tree, assume Left skewed, for calculating the height on left side would be O(n) and since there are no nodes on the right.  So you cannot disregard the early stopping like you  mentioned.

aryyamajana: Changed back the time complexity to O(n2) as Jaspinder sir suggested that O(n2) is the correct answer.

jaspindersingh83:
I made a mistake her @aryyamajana . @shrivatsav is absolutely...right here. Just saw the video explanation

aryyamajana:
> I made a mistake her @aryyamajana . @shrivatsav is absolutely...right here. Just saw the video explanation

Ok sir, thanks for the update.

PR #514, opened 2022-03-21
Competitive-Coding-3 completed - Please review

simransarin:


keshav-Gupta-IS (commented):
Good Job, Simran!! Keep it up. Also keep in mind although approach is two pointers when in interviews do say references since Java does not have pointers.

keshav-Gupta-IS:
Good Job, Simran!! Keep it up. Also keep in mind although approach is two pointers when in interviews do say references since Java does not have pointers.

simransarin:
wow, I never noticed. That is great feedback. thanks for reading my comments so closely Keshav, appreciate it!

PR #559, opened 2022-05-16
tried competitive coding 3 problems

Charishma090619999:


FindPairsDiff.py
```diff
@@ -0,0 +1,20 @@
...
+            i+=1
+        if i==len(nums):
+            break
+        if nums[i]+target in dict1:
+            res.append((nums[i],nums[i]+target))
+        if nums[i]-target in dict1:
+            res.append((nums[i],nums[i]-target))
+        dict1[nums[i]]=1
+        i+=1
+    return res
```
Sumesh-H, line 18: The output should be an integer, why are you returning the array? Please redo your code and submit it for review.

Sumesh-H (commented):
Good job, please address the comments.

Sumesh-H:
Good job, please address the comments.

Sumesh-H:
> Hi , Do we need to return the count of pairs which give up difference of target? Thanks and regards, Charishma Anubrolu.
> […](#)

Yes, you need to return count of pairs whose absolute difference is K

PR #627, opened 2022-07-15
Completed Competitive_Coding-3 - Please Review

sahithi6445:


k-Diff.java
```diff
@@ -0,0 +1,44 @@
+//Time complexity: O(N logN)
+//Space complexity: O(1)
```
daredevilteja, line 2: SC is O(n) because if you use mergesort or quicksort for sorting then SC is O(n)

pascalTriangle.java
```diff
@@ -0,0 +1,42 @@
+// Time Complexity : O(nk)->for rows and cols
+// Space Complexity : O(k)->for columns
```
daredevilteja, line 2: O(1)

daredevilteja (commented):
Please check the comments

daredevilteja:
Good Work! Keep it up

PR #630, opened 2022-07-18
completed CC-3 Please Check

tusharpawar611:


niharikagupta2021 (commented):
Good Job! please add sc and tc as well. Complete the other problem too

niharikagupta2021:
Good Job! please add sc and tc as well. Complete the other problem too

PR #762, opened 2023-07-15
k-diff pairs problem solved

Erlybird:


poojaGit95 (commented):
Good job. Briefly explain the approach used

poojaGit95:
Good job. Briefly explain the approach used

PR #767, opened 2023-07-18
Complete please review

parthvk:


p1.java
```diff
@@ -0,0 +1,37 @@
+// Time Complexity : O(n)
```
keshav-Gupta-IS, line 1: Check oce more it should be n^2

keshav-Gupta-IS (commented):
Good Job, Path!! Keep it up.

keshav-Gupta-IS:
Good Job, Path!! Keep it up.

p1.java
parthvk: Yes it should be n^2. I will correct it. Thanks!

PR #794, opened 2023-10-10
Completed K-diff pairs problem

rinkyfulwani:


pvk3081 (commented):
Good Job Rinky. Please add the other problem too

pvk3081:
Good Job Rinky. Please add the other problem too

PR #848, opened 2024-01-20
Completed CC-3

amishpapneja:


118-pascals-triangle.py
```diff
@@ -0,0 +1,56 @@
...
+1 <= numRows <= 30
+
+
+"""
+
+class Solution:
+    """
+    ACcepted
+    Time Complexity: O(n^2)
+    Space Complexity: O(n^2)
```
ishika2741, line 33: Check space complexity

ishika2741 (commented):
Good work!

ishika2741:
Good work!

118-pascals-triangle.py
amishpapneja: Are you suggesting that the nested list to be returned should not be counted as auxiliary memory? And its O(1)

PR #946, opened 2024-11-22

Zubin-glitch:


Nandha2Tech (commented):
Great Work! The logic of having two pointers fails in finding max subarray sum since you are considering that the window shrinks from both the edges based on the checks. But in this scenario arr = [4, -1, 2, 1, -7, 5] low ptr moves away from 4 based on the checks but the max subarray is [4,-1,2,1] which is missed.

Nandha2Tech:
Great Work! The logic of having two pointers fails in finding max subarray sum since you are considering that the window shrinks from both the edges based on the checks. But in this scenario arr = [4, -1, 2, 1, -7, 5] low ptr moves away from 4 based on the checks but the max subarray is [4,-1,2,1] which is missed.

PR #1116, opened 2025-08-11
Done Competitive-Coding-1

akhilsai25:


super30admin:
- PascalTriangle.java: Excellent job! The solution is correct, efficient, and well-structured. The comments help in understanding the logic, and the code follows best practices.
- PairsToK.java: The solution is correct and efficient, but consider using a more efficient data structure for storing pairs, such as a Set<String> where each pair is stored as a concatenated string (e.g., "a,b"). This could reduce the overhead of List operations and make the code slightly more efficient. Additionally, the variable name 'map' is misleading since it's a Set, not a Map. Consider renaming it to 'seen' or 'numbers' for clarity.

PR #1139, opened 2025-09-16
Completed both the problems

restacksyj:


super30admin:
It seems there might have been an error in your submission process as the file contains binary data rather than code. Please ensure you're submitting the correct file with your solution code. For the Pascal's Triangle problem, the reference solution shows a clear approach using nested loops to build each row based on the previous one. When resubmitting, make sure your solution:
- Correctly implements the Pascal's Triangle generation
- Has proper comments explaining the logic
- Follows good coding practices (proper indentation, meaningful variable names)
- Handles edge cases (like numRows = 0 or 1)

PR #1140, opened 2025-09-16
Completed both problems

restacksyj:


super30admin:
- Strengths:
  - The logic is straightforward and easy to follow.

- Areas for Improvement:
  - For Exercise2.cpp, the time complexity should be corrected to O(n log n) due to the sorting step.
  - Avoid using `#include <bits/stdc++.h>` in favor of including only necessary headers.
  - Consider adding more edge case checks, such as handling empty input or negative values of `k` in Exercise2.cpp.

PR #1143, opened 2025-10-18
done competitive coding3

ukvaddi:


super30admin:
- For problem1.cs: Well done! Your solution is correct, efficient, and clean. The logic is clear and matches the expected approach for generating Pascal's Triangle.
- For problem2.cs: Your approach is on the right track, but there are some inefficiencies. Instead of using strings to represent pairs, consider using a HashSet of tuples or a more efficient data structure. Also, ensure that you handle duplicate pairs correctly without relying on string operations. The current implementation may miss some edge cases, such as when there are duplicate numbers in the array.

PR #1144, opened 2025-10-20
Competitive_Coding-3 Completed

ashishbadhe97:


super30admin:
- For KDiffPairs.java: Well done! The solution is efficient and correctly handles the edge case where k=0. The use of a hash map to count frequencies is a good approach.
- For PascalsTriangle.java: The solution has a logical error. The loop should start from i=0 or i=1 to include the first two rows of Pascal's Triangle. Additionally, the condition `if(j == 0 || j == i)` is correct, but the implementation misses the base cases. You should initialize the first row explicitly before the loop. Also, the variable names could be more descriptive (e.g., `prevRow` instead of `prevList`).

PR #1163, opened 2026-01-13
completed competitive coding 3

abhayb94:


super30admin:
- Problem1.java: The solution is excellent. The approach is clear, and the implementation is correct. No improvements are needed.
- Problem2.java: The solution is mostly correct, but there is a minor issue with the initial check for k < 0. Instead of returning -1, it should return 0, as no pairs can exist with a negative k. Additionally, the use of a HashSet of Lists is correct but could be optimized further by using a HashSet of Strings or a custom Pair class to reduce overhead.

PR #1170, opened 2026-02-22
Add implementation for leetcode problems from mock interview.

rishigoswamy:


super30admin:
Your solution for Pascal's Triangle is excellent. It is efficient, correct, and well-commented. The approach of initializing each row with 1s and then filling the middle values using the previous row is optimal. The code is clean and easy to understand.

One minor suggestion: In the Pascal's Triangle solution, you can consider adding a check for numRows=0 at the beginning to return an empty list, although the problem states that numRows is at least 1, so it's not strictly necessary.

For the K-diff Pairs solution, it is also correct and efficient. However, note that in the case k>0, you are iterating over all keys in the hashmap, which includes duplicates. But since you are using the keys (which are unique), it is correct. Alternatively, you could iterate over the unique keys by using a set, but your current approach is already efficient.

Overall, great job on both solutions!

PR #1173, opened 2026-02-23
Coding-3

nagasai67:


super30admin:
Your solution for Pascal's Triangle is correct and efficient. Well done! Here are a few suggestions to improve code clarity:
- Consider renaming "dup_list" to something more descriptive like "current_row" or "new_row" to make the code easier to understand.
- You might want to add a check for numRows=0 at the beginning, though the problem states numRows>=1, so it's not strictly necessary.
- The inner loop range is correct, but you can also write it as `for j in range(1, i):` since the inner elements are from index 1 to i-1 (for a row of length i+1). This might be slightly more intuitive.

Overall, your solution is solid and meets all requirements.

PR #1177, opened 2026-04-14
Complete competitive-coding-3

paridhimalviya:
-> Pascal's triange
-> Pairs with k difference

### DIFF
--- a/PairsWithKDiff.swift
+++ b/PairsWithKDiff.swift
@@ -1,0 +1,53 @@
+//
+//  PairsWithKDiff.swift
+//  DSA-Practice
+//
+//  Created by Paridhi Malviya on 4/5/26.
+//
+
+class PairsWithKDiff {
+    
+    init() {
+        
+    }
+    
+    func findPairs(_ nums: [Int], _ k: Int) -> Int {
+        if (k < 0) {
+            return 0
+        }
+
+        var seen = Set<Int>()
+        //store the smaller element in result to avoid duplicates
+        var result = Set<Int>()
+        for num in nums {
+             if (seen.contains(num - k)) {
+                result.insert(num-k)
+             }
+             if (seen.contains(num + k)) {
+                result.insert(num)
+             }
+             seen.insert(num)
+        }
+        return result.count
+    }
+    
+    func findPairsUsingMap(_ nums: [Int], _ k: Int) -> Int {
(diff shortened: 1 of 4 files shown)


super30admin:
### Interview Problem : Pascal's Triangle (PascalsTriangle.swift)
Your solution is correct and handles the problem as required. Here are some suggestions for improvement:

1. **Code Structure**: Instead of handling rows 0 and 1 separately, you can initialize the first row and then use a loop to generate the remaining rows. This reduces redundancy and makes the code more concise.

2. **Initialization**: Instead of creating an array with placeholder values (-1), consider building the row by appending elements. This avoids initializing with unnecessary values and is more idiomatic.

3. **Variable Names**: Use more descriptive variable names. For example, `output` is good, but `o` for a row is not clear. Consider using `row` instead.

4. **Efficiency**: The current approach is efficient, but you can avoid the special case for the first two rows by starting with the first row and then generating each subsequent row based on the previous one.

Here's a revised version of your code for reference:
```swift
func generate(_ numRows: Int) -> [[Int]] {
    guard numRows > 0 else { return [] }
    var result = [[1]]
    for i in 1..<numRows {
        var row = [1]
        let prevRow = result[i-1]
        for j in 1..<i {
            row.append(prevRow[j-1] + prevRow[j])
        }
        row.append(1)
        result.append(row)
    }
    return result
}
```

This version is more streamlined and handles all rows uniformly.

VERDICT: PASS

---

### Interview Problem: Pairs with K difference  (PairsWithKDiff.swift)
Your solution `findPairsUsingMap` is correct and efficient. It handles both k=0 and k>0 cases appropriately. The time and space complexity are O(n), which is optimal.

However, your first solution `findPairs` has a syntax error: there is an extra closing parenthesis in the line `if (seen.contains(num + k))`. This would cause a compilation error. If you fix that, the algorithm should work correctly. But note that this method might be less efficient because it uses a set for the result and performs two checks per element. Also, it might be less readable than the map-based solution.

For the problem constraints, the map-based solution is preferable because it is straightforward and avoids potential issues with ordering during traversal. 

I recommend using the `findPairsUsingMap` method. You can also consider improving the first method by fixing the syntax and perhaps adding comments to explain why storing the smaller element avoids duplicates.

VERDICT: PASS

PR #1184, opened 2026-05-10
add interview-3 solutions

nikhylw:
Solution 1 (k-diff pairs) - W3_interview_k-diff-pairs.py
Solution 2 (pascals) - W3_interview_pascals.py


super30admin:
### Interview Problem : Pascal's Triangle (W3_interview_pascals.py)
**Strengths:**
- Clear, methodical approach with three distinct phases
- Good use of descriptive comments
- Clean variable naming
- Correctly handles edge cases (numRows=1 works correctly)

**Areas for Improvement:**
- The solution could be slightly more efficient by directly building rows instead of initializing with zeros and then overwriting. For example, directly pushing {1} for first row, then building subsequent rows by adding elements.
- The comment "Insert 0s for all the rows in increasing order" is slightly misleading - you're not inserting zeros, you're creating rows of zeros.
- Could add type hints for better code documentation (e.g., `def generate(self, numRows: int) -> List[List[int]]:`)

**Minor Optimization Opportunity:**
The reference solution is more space-efficient during construction as it builds rows incrementally. Your approach pre-allocates all rows with zeros, which is fine but slightly less elegant. However, both approaches have the same final space complexity.

VERDICT: PASS

---

### Interview Problem: Pairs with K difference  (W3_interview_k-diff-pairs.py)
**Strengths:**
- Excellent single-pass solution that is more efficient than the reference
- Clean, readable code with good variable naming
- Correctly handles all edge cases including k=0
- Good use of sets for tracking seen numbers and unique pairs
- Well-commented with time/space complexity

**Areas for Improvement:**
- The solution is already well-optimized. One minor consideration: for very large arrays with small k values, the approach is already near-optimal.
- Could consider adding type hints for better code documentation (e.g., `nums: List[int], k: int` already has them, which is good)

The solution demonstrates a solid understanding of the problem and implements an efficient algorithm. The single-pass approach is superior to the reference solution's two-pass method.

VERDICT: PASS

